PR #6, opened 2022-07-18
Solution

YuriiKindrat:


cinema/serializers.py
```diff
@@ -25,7 +25,7 @@ class Meta:
 class MovieSerializer(serializers.ModelSerializer):
     class Meta:
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = ("id", "title", "description", "duration", "genres", "actors", "image")
```
mens0, line 28: if you add "image" field here, then you will be able to add image on the movies list page. According to the taks, you shouldn't be able to do that

mens0 (changes requested):

MasterpieceElbow (approved):

PR #10, opened 2022-07-19
solution

DmytroSvirsa:


cinema/serializers.py
```diff
@@ -52,7 +88,11 @@ class Meta:
 
 class MovieSessionListSerializer(MovieSessionSerializer):
     movie_title = serializers.CharField(source="movie.title", read_only=True)
-    cinema_hall_name = serializers.CharField(source="cinema_hall.name", read_only=True)
+    movie_image = serializers.CharField(source="movie.image", read_only=True)
```
Serhii-Leonenko, line 91: you can use `serializer.ImageField`

MasterpieceElbow, line 91: Serhii is right. You are supposed to use `serializers.ImageField`

MasterpieceElbow (approved):

cinema/serializers.py
MasterpieceElbow, line 91: Serhii is right. You are supposed to use `serializers.ImageField`

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #19, opened 2022-07-19
Solution

Kiawami:


daria-niklaevna:
You should in py-images-and-custom-auth-serializer in tests  test_movie_api.py replace 
        self.user = get_user_model().objects.create_superuser(
            'admin', 'admin@myproject.com', 'password'
        )
to 
        self.user = get_user_model().objects.create_superuser(
            'admin@myproject.com', 'password'
        )
On line 57 of the code.

PR #20, opened 2022-07-19
Solution

vasylhnatiuk:


cinema/admin.py
```diff
@@ -1,6 +1,7 @@
 from django.contrib import admin
 
-from .models import CinemaHall, Genre, Actor, Movie, MovieSession, Order, Ticket
+from .models import CinemaHall, Genre, Actor,\
```
KatiukhaO, line 3: does not correct format to write import. Use (...)

cinema/views.py
```diff
@@ -22,6 +24,7 @@
     MovieListSerializer,
     OrderSerializer,
     OrderListSerializer,
+    MovieImageSerializer,
```
KatiukhaO, line 27: you should have similar format write import

KatiukhaO (changes requested):

MasterpieceElbow (changes requested):
Your tests have failed. It seems that you haven't pushed user migrations.

PR #31, opened 2022-07-19
Solution

darten360:


cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
KatiukhaO, line 26: will be better if to have similar format import, do import MovieImageSerializer, with new line

MasterpieceElbow (approved):

PR #49, opened 2022-09-16
Solution

DHushchyk:


IvanRamyk (changes requested):
An image shouldn't be on PR

IvanRamyk (approved):
Well done!

PR #72, opened 2022-11-22
Solution

olia-trofymchuk:


cinema/models.py
```diff
@@ -12,40 +16,49 @@ class CinemaHall(models.Model):
...
     @property
-    def full_name(self):
+    def full_name(self) -> str:
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(movie, filename):
+    _, ext = os.path.splitext(filename)
+
+    filename = f"{slugify(movie.title)}-{uuid.uuid4()}{ext}"
```
mltdll, line 45: I think it's better in terms of readability to avoid calling functions or performing operations inside an f-string. But it's up to you to decide if you want to change it, I guess.

cinema/serializers.py
```diff
@@ -33,7 +33,16 @@ class Meta:
...
+        fields = (
+            "id",
+            "title",
+            "description",
+            "image",
+            "duration",
+            "genres",
+            "actors",
+        )
+        read_only_fields = ("image",)
```
mltdll, line 45: Shouldn't `"id"` also be specified here?

mltdll (changes requested):
By the way, cool Inception poster! But it probably wasn't really necessary to include it in your PR. :)

cinema/serializers.py
olia-trofymchuk: yees😄 
my fault
thank youuu)

mltdll (approved):

cinema/models.py
```diff
@@ -12,40 +16,51 @@ class CinemaHall(models.Model):
...
         return self.first_name + " " + self.last_name
 
     @property
-    def full_name(self):
+    def full_name(self) -> str:
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(movie, filename):
+    _, ext = os.path.splitext(filename)
```
Y-Havryliv, line 43: `extension` would be better

Y-Havryliv (changes requested):

PR #76, opened 2022-11-23
Solution

KatrinLazarenko:


pavlo-kapitanets (changes requested):
I think it's better to add images from fixture to .gitignore, because we use it only for debagging our code

pavlo-kapitanets (approved):
GJ!

MasterpieceElbow (approved):
GJ!

PR #78, opened 2022-11-23
Solution

Kitazuka:


sbublinskyi (approved):

TPolina (changes requested):
you don't redefine UserAdmin in admin.py so that username was not obligatory

PR #80, opened 2022-11-24
Created models, serializers for image upload

alexzubkoff:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
mltdll, line 44: In my opinion, it's better in terms of readability to avoid calling functions or performing operations inside an f-string. But it's up to you, I guess.

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
mltdll, line 26: Don't forget to fix the styling in your imports. :)

mltdll (changes requested):

PR #84, opened 2022-11-24
Solution

mltdll:


AVyha (approved):
Looks good!

cinema/serializers.py
```diff
@@ -49,9 +58,13 @@ class MovieDetailSerializer(MovieSerializer):
     genres = GenreSerializer(many=True, read_only=True)
     actors = ActorSerializer(many=True, read_only=True)
 
+
+class MovieImageSerializer(MovieSerializer):
+    image = serializers.ImageField(required=True)
```
Nattalli, line 63: btw, `required=True` is a default value, so that is unnecessary here

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #89, opened 2022-11-24
solution

AVyha:


mltdll (changes requested):
The solution seems good, but I guess it's better to remove the images from the PR. :)

cinema_service/settings.py
```diff
@@ -121,7 +121,7 @@
 
 LANGUAGE_CODE = "en-us"
 
-TIME_ZONE = "UTC"
+TIME_ZONE = False
```
mltdll, line 124: Oops, missed it in the initial review. Maybe you meant `USE_TZ = False`? :)

mltdll (changes requested):

cinema_service/settings.py
AVyha: 😊

mltdll (approved):
GJ!

Nattalli (approved):
👍🏻

PR #96, opened 2022-11-26
Solution

fsocie7y:


cinema/serializers.py
```diff
@@ -44,14 +44,31 @@ class MovieListSerializer(MovieSerializer):
         many=True, read_only=True, slug_field="full_name"
     )
 
+    class Meta(MovieSerializer.Meta):
+        fields = MovieSerializer.Meta.fields + ("image", )
```
LiudmylaKulinchenko, line 48: maybe you can just add field image in base MovieSerializer

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
LiudmylaKulinchenko, line 26: move this import to the next line

cinema/views.py
```diff
@@ -101,8 +103,26 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser, ]
```
LiudmylaKulinchenko, line 115: I think you don't need permission_classes here

LiudmylaKulinchenko (changes requested):

cinema/serializers.py
fsocie7y: no, then test didn't pass, because we shuld can add image only on detail page

cinema/views.py
fsocie7y: i think its real important, because no-admin user can't add image to movie

LiudmylaKulinchenko: you already have such permission in this view

LiudmylaKulinchenko (approved):

cinema/views.py
TPolina: I also think that you don't need permission_classes here

TPolina (approved):

PR #103, opened 2022-11-30
Solution. Serializer implemented.

AlenOl:
Added ImageField image to the Movie model.
Replaced username field with email field in User model.
Wrote custom AuthTokenSerializer.


IvanRamyk (approved):
Well done!

PR #125, opened 2023-02-02
Solution

ansicat:


user/serializers.py
```diff
@@ -22,3 +23,38 @@ def update(self, instance, validated_data):
...
+        if email and password:
+            user = authenticate(
+                request=self.context.get("request"),
+                email=email,
+                password=password,
+            )
+
+            # The authenticate call simply returns None for is_active=False
+            # users. (Assuming the default ModelBackend authentication
+            # backend.)
```
andrewbelo: Коменти прибери

cinema/views.py
```diff
@@ -101,8 +103,22 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=self.request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
andrewbelo: https://www.django-rest-framework.org/api-guide/serializers/#validation
The .is_valid() method takes an optional raise_exception flag that will cause it to raise a serializers.ValidationError exception if there are validation errors.

These exceptions are automatically dealt with by the default exception handler that REST framework provides, and will return HTTP 400 Bad Request responses by default.

andrewbelo (changes requested):

cinema/serializers.py
```diff
@@ -44,14 +50,34 @@ class MovieListSerializer(MovieSerializer):
...
+            "actors",
+            "image",
+        )
+
 
 class MovieDetailSerializer(MovieSerializer):
     genres = GenreSerializer(many=True, read_only=True)
     actors = ActorSerializer(many=True, read_only=True)
 
     class Meta:
```
ArsenPidhoretskyi, line 70: try:
```
class Meta(MovieSerializer.Meta):
    pass
```

Or try without meta

user/serializers.py
```diff
@@ -22,3 +23,35 @@ def update(self, instance, validated_data):
...
+
+class EmailAuthTokenSerializer(serializers.Serializer):
+    email = serializers.EmailField(label=_("Email address"), write_only=True)
+    password = serializers.CharField(
+        label=_("Password"),
+        style={"input_type": "password"},
+        trim_whitespace=False,
+        write_only=True,
+    )
+    token = serializers.CharField(label=_("Token"), read_only=True)
```
ArsenPidhoretskyi: You do nit use this as I can see)

ArsenPidhoretskyi (changes requested):

cinema/serializers.py
ansicat: Class MovieListSerializer.Meta was redefined in order to add field "image" because task description requires to show one in Movie list endpoint. MovieSerializer.Meta doesn't have "image" in it's fields tuple.

Bohdan-Salamakha (approved):

PR #143, opened 2023-03-19
My Solution

yanpurdenko:


user/models.py
```diff
@@ -1,5 +1,54 @@
...
+        if not email:
+            raise ValueError("The given email must be set")
+        email = self.normalize_email(email)
+        user = self.model(email=email, **extra_fields)
+        user.set_password(password)
+        user.save(using=self._db)
+        return user
+
+    def create_user(
+            self, email: str, password: str = None, **extra_fields
```
Bohdan-Salamakha: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contain string elements and can be None has typehint Optional[list[str]]

user/models.py
```diff
@@ -1,5 +1,54 @@
...
+    def create_superuser(
+            self, email: str, password: str, **extra_fields
+    ) -> ValueError | User:
+        """Create and save a SuperUser with the given email and password."""
+        extra_fields.setdefault("is_staff", True)
+        extra_fields.setdefault("is_superuser", True)
+
+        if extra_fields.get("is_staff") is not True:
+            raise ValueError("Superuser must have is_staff=True.")
+        if extra_fields.get("is_superuser") is not True:
```
Bohdan-Salamakha: use "is" only to compare value with None

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (approved):
Well done!

PR #147, opened 2023-03-23
Solution

derkach55:


Grayder0152 (changes requested):
image
Pls, resolve test issue

cinema/models.py
```diff
@@ -35,12 +39,22 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_custom_path(instance: object, filename: str) -> str:
+    _, ext = os.path.splitext(filename)
+    filename = os.path.join(
+        "uploads/images/",
+        f"{slugify(instance.title)}-{uuid.uuid4()}.{ext}"
```

cinema/models.py
```diff
@@ -35,12 +39,22 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_custom_path(instance: object, filename: str) -> str:
```
ArsenPidhoretskyi: instance
Please provide the correct type, you can specify more better type

ArsenPidhoretskyi (changes requested):

cinema/models.py
derkach55: how?
Movie class is not defined yet

vasylhnatiuk: you can use "" 

`def create_custom_path(instance: "Movie", filename: str) -> str:`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job!

PR #153, opened 2023-04-05
Develop

artur-leleiko:


cinema/views.py
```diff
@@ -101,8 +103,20 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="movie-upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
```
olena-hul: No need to use if there. just `serializer.is_valid(raise_exception=True)`

olena-hul (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_custom_path(instance, filename: str) -> str:
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+    return os.path.join("uploads/images/", filename)
```
ArsenPidhoretskyi: due to task requirements:
uploads/movies/

ArsenPidhoretskyi: Also i recoment to do as:
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #156, opened 2023-04-06
1 commit

AndereLion:


manjustice (approved):

cinema/views.py
```diff
@@ -101,18 +103,43 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        detail=True,
+        methods=["POST"],
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
```
wQuelS: You have already defined IsAdminOrIfAuthenticatedReadOnly permission class. And it will work here as indended

user/serializers.py
```diff
@@ -10,15 +11,44 @@ class Meta:
...
+                    msg = _("User account is disabled.")
+                    raise serializers.ValidationError(
+                        msg, code="authorization"
+                    )
+            else:
+                msg = _("Unable to log in with provided credentials.")
+                raise serializers.ValidationError(msg, code="authorization")
+        else:
+            msg = _("Must include 'username' and 'password'.")
+            raise serializers.ValidationError(msg, code="authorization")
```
wQuelS: this can be simplified if you start from `if not user`

wQuelS (changes requested):

user/serializers.py
AndereLion: Fixed all )

wQuelS (approved):

PR #158, opened 2023-04-07
Dev

SafonovVladimir:
add solution


cinema/views.py
```diff
@@ -92,34 +97,55 @@ def get_queryset(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
```
wQuelS: You have already defined IsAdminOrIfAuthenticatedReadOnly permission class. So it will work in this case as intended

wQuelS (changes requested):

cinema/views.py
SafonovVladimir: Ok. Fixed

Y-Havryliv (approved):
Well done!

PR #162, opened 2023-04-09
Solution

SergiiMachulin:


user/serializers.py
```diff
@@ -10,15 +11,48 @@ class Meta:
...
+    password = serializers.CharField(
+        label=_("Password"),
+        style={"input_type": "password"},
+        trim_whitespace=False,
+        write_only=True
+    )
+    token = serializers.CharField(
+        label=_("Token"),
+        read_only=True
+    )
```
yura-hudzovskyi, line 40: if you don`t use it in validation, you can delete it

yura-hudzovskyi (changes requested):

user/serializers.py
SergiiMachulin: Not agree. It doesn't depend on validation method including. 
The token field in the AuthTokenSerializer class is set by default and also set as a read_only field. It was there when we used username, but we didn't override AuthTokenSerializer but it has token field) 

However, including the token field in the serializer can be useful to provide a consistent response format to the client, and to make it clear that the authentication was successful and a token was generated.

Additionally, including the token field in the serializer class can make it easier to test the authentication process, as the generated token can be easily accessed in the test code without having to parse the response body.

So, while it is not strictly necessary to include the token field in the AuthTokenSerializer class, it is generally considered good practice to do so.

cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
```
wQuelS: You have already defined IsAdminOrIfAuthenticatedReadOnly permission class, so this is redundant here

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #165, opened 2023-04-10
Solution DonchykD

DmytryTong:


bihunva (changes requested):
Looks good but I think there is no need to upload media files to github)

DmytryTong:
Images deleted

bihunva (approved):
GJ

cinema/views.py
```diff
@@ -101,8 +104,29 @@ def get_serializer_class(self):
...
+
         return MovieSerializer
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[
+            IsAdminUser,
+        ]
```
Serhii-Leonenko, line 118: You have already defined permission_class IsAdminOrIfAuthenticatedReadOnly, it is working properly for this case

user/serializers.py
```diff
@@ -22,3 +26,29 @@ def update(self, instance, validated_data):
...
+        if email and password:
+            try:
+                user_obj = User.objects.get(email=email)
+                if not user_obj.check_password(password):
+                    raise serializers.ValidationError("Incorrect credentials")
+                data["user"] = user_obj
+            except User.DoesNotExist:
+                raise serializers.ValidationError(
+                    "This email is not registered"
+                )
```
Serhii-Leonenko, line 51: it's better to remove the exception handler and use authenticate() method for user, also it will check that user is_active

Serhii-Leonenko (changes requested):

PR #169, opened 2023-04-10
Solution

bihunva:


amber-marichi (changes requested):
very nice, but i think you better delete  .json  fixture file from your commit, also maybe remove comments from user/models.py just to make it more according to checklist ;)

bihunva:
> very nice, but i think you better delete .json fixture file from your commit, also maybe remove comments from user/models.py just to make it more according to checklist ;)

oops, forgot about json file, thanks. there are no comments in the user/models.py, but docstrings

amber-marichi (approved):
ok then. goodjob!

YuriiKindrat (approved):

PR #174, opened 2023-04-11
Solution

VladyslavCherkashyn:


DavidGlusch (approved):

user/serializers.py
```diff
@@ -22,3 +23,26 @@ def update(self, instance, validated_data):
...
+        password = attrs.get("password")
+
+        user = authenticate(request=self.context.get("request"),
+                            username=email, password=password)
+
+        if not user:
+            msg = "Unable to authenticate with provided credentials"
+            raise serializers.ValidationError(msg, code="authentication")
+
+        token, _ = Token.objects.get_or_create(user=user)
```
YuriiKindrat: why do u need this ?

user/serializers.py
```diff
@@ -22,3 +23,26 @@ def update(self, instance, validated_data):
...
+    def validate(self, attrs):
+        email = attrs.get("email")
+        password = attrs.get("password")
+
+        user = authenticate(request=self.context.get("request"),
+                            username=email, password=password)
+
+        if not user:
+            msg = "Unable to authenticate with provided credentials"
+            raise serializers.ValidationError(msg, code="authentication")
```
YuriiKindrat, line 41: what if user is not active ?

YuriiKindrat (changes requested):

wQuelS (approved):

PR #179, opened 2023-04-11
solutionis

skmiwa1:


viktoria-rybenchuk (approved):
GJ!

user/serializers.py
```diff
@@ -22,3 +23,33 @@ def update(self, instance, validated_data):
...
+                    msg = _("User account is disabled.")
+                    raise serializers.ValidationError(
+                        msg, code="authorization"
+                    )
+            else:
+                msg = _("Unable to log in with provided credentials.")
+                raise serializers.ValidationError(msg, code="authorization")
+        else:
+            msg = _("Must include 'username' and 'password'.")
+            raise serializers.ValidationError(msg, code="authorization")
```
wQuelS: this can be simplified a bit:)
start from `if not user`

wQuelS (changes requested):

user/serializers.py
skmiwa1: got it!

wQuelS (approved):
Really good job

PR #193, opened 2023-04-11
Solution

myroslavatrschk:


YuriiKindrat (changes requested):
all tests should pass

user/serializers.py
```diff
@@ -22,3 +23,37 @@ def update(self, instance, validated_data):
...
+    )
+
+    def validate(self, attrs):
+        email = attrs.get("email")
+        password = attrs.get("password")
+
+        if email and password:
+            user = authenticate(
+                request=self.context.get("request"),
+                username=email,
```
Y-Havryliv: Replace 'username=email' with 'email=email' in the authenticate function, since we are using email field instead of username in the User model.

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
Y-Havryliv: won't you have an extra dot here?

Y-Havryliv (changes requested):

Y-Havryliv (changes requested):
tests failed

Y-Havryliv (approved):
Nice!

PR #197, opened 2023-04-11
Solution

Anatolii-Poznyak:


user/serializers.py
```diff
@@ -22,3 +23,28 @@ def update(self, instance, validated_data):
...
+        write_only=True,
+        label=_("Password")
+    )
+
+    def validate(self, attrs):
+        email = attrs.get("email")
+        password = attrs.get("password")
+        user = authenticate(request=self.context.get("request"),
+                            email=email,
+                            password=password)
```
YuriiKindrat: it is better to check email/password existence before authentication and raise validation error if not

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #203, opened 2023-04-12
'Solution'

NickSavchyn:


user/serializers.py
```diff
@@ -22,3 +24,32 @@ def update(self, instance, validated_data):
...
+        label=_("Password"),
+        style={"input_type": "password"},
+        trim_whitespace=False,
+        write_only=True
+    )
+
+    def validate(self, attrs):
+
+        email = attrs.get("email")
+        password = attrs.get("password")
```
YuriiKindrat, line 41: it is better to check email/password existence before auth, and raise validation error if not

YuriiKindrat (changes requested):
all tests should pass

YuriiKindrat (approved):

PR #204, opened 2023-04-13
Solution

ykopatko:


Serhii-Leonenko (changes requested):
All checks should be passed

ykopatko:
> All checks should be passed

sure, it is done now!

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
YuriiKindrat: fix indentation here

YuriiKindrat (changes requested):

cinema/views.py
ykopatko: done

YuriiKindrat (approved):

PR #206, opened 2023-04-15
Solution

anastasiia-tsurkan:


cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
olena-hul: The `if` is not needed, you can simply call `serializer.is_valid(raise_exception=True)`

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #213, opened 2023-05-06
Solution 

Oleksiy-Liubchenko:


.gitignore
```diff
@@ -8,3 +8,4 @@ venv/
 **__pycache__/
 *.pyc
 **db.sqlite3
+media
```
ArsenPidhoretskyi: Add empty line

cinema/views.py
```diff
@@ -68,6 +71,21 @@ class MovieViewSet(
...
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi: instead of if-else use serializer.is_valid(raise_exception=True)

ArsenPidhoretskyi (changes requested):

cinema/views.py
```diff
@@ -68,6 +71,20 @@ class MovieViewSet(
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
```
ArsenPidhoretskyi: if is not needed
```suggestion
        serializer.is_valid(raise_exception=True):
```

ArsenPidhoretskyi (changes requested):

cinema/views.py
Oleksiy-Liubchenko: oops

ArsenPidhoretskyi (approved):
well done!

PR #214, opened 2023-05-07
Develop

U-Ra-All:


cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi: Use serializer.is_valid(raise_exception=True) instead of if-else

cinema/urls.py
```diff
@@ -18,6 +18,8 @@
 router.register("movie_sessions", MovieSessionViewSet)
 router.register("orders", OrderViewSet)
 
-urlpatterns = [path("", include(router.urls))]
+urlpatterns = [
+    path("", include(router.urls))
+]
```
ArsenPidhoretskyi: No need to include if basepath is not provided
If you change this change to 
urlpatterns = router.urls

ArsenPidhoretskyi (changes requested):

cinema/views.py
U-Ra-All: Fixed

cinema/urls.py
U-Ra-All: Fixed

olena-hul (approved):

olena-hul:
Good job!

U-Ra-All:
> Good job!

Thanks!

PR #222, opened 2023-05-23
'py-images-and-custom-auth-serializer'

OmeLchDO:


cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: use serializer.is_valid(raise_exception=True) instead of if-statement

cinema/serializers.py
```diff
@@ -15,58 +15,119 @@
...
-    genres = GenreSerializer(many=True, read_only=True)
-    actors = ActorSerializer(many=True, read_only=True)
+    genres = GenreSerializer(
+        many=True,
+        read_only=True
+    )
+    actors = ActorSerializer(
+        many=True,
+        read_only=True
+    )
```
ArsenPidhoretskyi, line 84: Image must present on list and detail

ArsenPidhoretskyi (changes requested):
Tests are failed: test_image_url_is_shown_on_movie_list, test_image_url_is_shown_on_movie_detail

cinema/serializers.py
OmeLchDO: fixed

cinema/views.py
OmeLchDO: fixed

vsmutok (approved):
Good 👍

PR #224, opened 2023-06-03
Solution

kostya-kononenko:


cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
 
+    @action(methods=["POST"],
+            detail=True,
+            url_path="upload-image",
+            permission_classes=(IsAdminUser,))
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use serializer.is_valid(raise_exception=True) instead of if-else

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good 👍

PR #229, opened 2023-06-07
image_and_custom_auth_serializer

VKungin:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(movie, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(movie.title)}-{uuid.uuid4()}.{extension}"
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "images", filename)
```

vsmutok: It must works in all OS

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #248, opened 2023-07-04
Solution v_1.0

bohdan-yatsyna:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
ArsenPidhoretskyi: Is really dot needed?

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
...
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
resolve comments

cinema/models.py
bohdan-yatsyna: You are right.

cinema/models.py
bohdan-yatsyna: Thank you for the remarks.

cinema/models.py
ArsenPidhoretskyi:
```suggestion
    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
```

cinema/models.py
ArsenPidhoretskyi: Not fixed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

cinema/models.py
bohdan-yatsyna: My bad, sorry.

cinema/models.py
bohdan-yatsyna: Now is fixed, thank you.

vasylhnatiuk (approved):

PR #249, opened 2023-07-05
Solution

abramov-petro:


cinema/models.py
```diff
@@ -35,12 +39,25 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+
+    return os.path.join("uploads/movies/", filename)
```
ArsenPidhoretskyi: Better to do like this, It will works in all OS:
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #250, opened 2023-07-06
Solution

DmytroBondariev:


cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: use serializer.is_valid(raise_exception=True)

vsmutok (changes requested):

.gitignore
```diff
@@ -7,4 +7,4 @@ venv/
 .pytest_cache/
 **__pycache__/
 *.pyc
-**db.sqlite3
+**db.sqlite3
```
ArsenPidhoretskyi: Restore this changes

cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
ArsenPidhoretskyi: Is dot really needed? Did you test manually?
```suggestion
    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
```

cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
...
+
+    return os.path.join("uploads/movies", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```
Better to do like this. It will works in all OS

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #251, opened 2023-07-07
solution

HalynaPetrova:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    return os.path.join(
+        "uploads/images/",
```
```suggestion
        "uploads", "images",
```
It must works in all OS

cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi: Use serializer.is_valid(raise_exception=True) instead of if-else statement

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Good job 👍

PR #254, opened 2023-07-12
solution

IvanGuculyak:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
+
+    return os.path.join("uploads/movies/", filename)
```
vsmutok: Better to do like this, It will works in all OS:
```suggestion
    return os.path.join("uploads", "movies", filename)
```

vsmutok (changes requested):
also remove movies from PR

vsmutok (approved):
Good 👍

PR #264, opened 2023-07-25
Develop

andy77andy:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_image_movie_path(instance, filename):
+    _, extensions = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extensions}"
```
Check files

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
...
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -94,13 +97,32 @@ def get_queryset(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload_image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        image = self.get_object()
+        serializer = self.get_serializer(image, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: use serializer.is_valid(raise_exception=True) instead of if statement

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #266, opened 2023-07-30
done

nazarbychkar:


cinema/views.py
```diff
@@ -101,8 +103,27 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
+    )
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: no need to use if statement. Instead use 

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #286, opened 2023-08-04
Solution

erikagayan:


requirements.txt
LiudmylaKulinchenko, line 1: make it text file, it is binary now

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
nvm

PR #302, opened 2023-08-14
Solution

Dimdimior:


cinema/views.py
```diff
@@ -101,8 +104,24 @@ def get_serializer_class(self):
...
+        detail=True,
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None) -> Response:
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev: Please, use `is_valid` with the argument `raise_exception=True` instead of `if-statement`.

```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #304, opened 2023-08-21
'Solution'

svitlana-savikhina:


cinema/views.py
```diff
@@ -94,13 +96,30 @@ def get_queryset(self):
...
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status.HTTP_200_OK)
```
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status.HTTP_200_OK)
```

EdAlekseiev: Use ` serializer.is_valid(raise_exception=True)` instead of if-statement

EdAlekseiev (changes requested):

cinema_service/settings.py
```diff
@@ -125,14 +125,17 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```
olena-hul, line 128: Remove media folder from the pull request

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #305, opened 2023-08-21
Solution

Niki-Alex:


cinema/models.py
```diff
@@ -1,6 +1,16 @@
...
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.conf import settings
+from django.utils.text import slugify
+
+
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+    return os.path.join("uploads/movies", filename)
```
```suggestion
   return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: No need to use if statement, instead use serializer.is_valid(raise_exception=True)

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #307, opened 2023-08-24
Solution

Oleksiy-Lyashenko:


cinema/models.py
```diff
@@ -35,12 +38,25 @@ def full_name(self):
...
 
 
+def slugify(title):
+    pass
+
+
+def movie_image_filename_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```

cinema/models.py
```diff
@@ -35,12 +38,25 @@ def full_name(self):
...
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -101,8 +103,22 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Instead of if-statement use serializer.is_valid(raise_exception=True)

ArsenPidhoretskyi (changes requested):

cinema/models.py
ArsenPidhoretskyi: Not fixed. Dot is not needed before {extension}

ArsenPidhoretskyi (changes requested):

pavlejviki (approved):
Good job!

PR #310, opened 2023-09-04
Solution

oleg-potichnyi:


cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
EdAlekseiev: You can use `is_valid` with `raise_exception=True` instead of `if`-statement

cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status.HTTP_400_BAD_REQUEST)
```
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

Bohdan-Salamakha (changes requested):
remove a media folder from github

Bohdan-Salamakha (approved):

PR #312, opened 2023-09-06
Solution

NataliiaPodoroha:


cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+
+    return os.path.join("uploads/movies/", filename)
```
vsmutok: it must work in all OS
```suggestion
  return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -101,8 +104,28 @@ def get_serializer_class(self):
...
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        """Endpoint for uploading movie image"""
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: Instead of if-statement use serializer.is_valid(raise_exception=True)

vsmutok (changes requested):

cinema/migrations/0002_movie_image.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.1 on 2023-09-06 12:18
+
+import cinema.models
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cinema', '0001_initial'),
```
vasylhnatiuk: change ' to ", use Black

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #316, opened 2023-09-24
Solution

Munkoor:


vsmutok (changes requested):
Please fix tests

cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok: You can use is_valid with raise_exception=True instead of if-statement

vsmutok (changes requested):
also please remove media files from PR

vsmutok (approved):
Great 👍

PR #320, opened 2023-10-05
Solution

nailia-mamiedova:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{ext}"
```
ArsenPidhoretskyi: Is dot really needed before {ext}

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
...
+
+    return os.path.join("upload/movies/", filename)
```
```suggestion
    return os.path.join("upload", "movies", filename)
```

cinema/views.py
```diff
@@ -101,8 +103,28 @@ def get_serializer_class(self):
...
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use raise_exception instead of if statement

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #323, opened 2023-10-12
Solution

Nagerba:


cinema/views.py
```diff
@@ -101,8 +103,20 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(detail=True, methods=["POST"], url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
```
EdAlekseiev: You don't need to use if-statement when you pass raise_exception=True to is_valid. It will raise exception automatically

cinema/views.py
```diff
@@ -101,8 +103,20 @@ def get_serializer_class(self):
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
ArsenPidhoretskyi: Is dot really needed before {extension}?

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
...
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
LGTM!

PR #330, opened 2023-10-16
Implemented task with additional images and using email instead of us…

ZaichenkoViktoriia:
…ername


Barsh4ec (approved):
Looks good, but I would remove media files from PR

cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
```
Serhii-Leonenko, line 127: you can use raise_exeption=True instead of these conditions

Serhii-Leonenko (approved):

PR #352, opened 2023-10-21
Solution

niksgawrilenko:


cinema/views.py
```diff
@@ -78,31 +80,44 @@ def get_queryset(self):
...
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
```
viktoria-rybenchuk: you can raise exception instead of this construction https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true

user/serializers.py
```diff
@@ -22,3 +23,38 @@ def update(self, instance, validated_data):
...
+        if email and password:
+            user = authenticate(
+                request=self.context.get("request"),
+                email=email,
+                password=password
+            )
+
+            # The authenticate call simply returns None for is_active=False
+            # users. (Assuming the default ModelBackend authentication
+            # backend.)
```
viktoria-rybenchuk: remove comment

viktoria-rybenchuk (changes requested):

user/serializers.py
niksgawrilenko: Done

cinema/views.py
niksgawrilenko: Done

PR #353, opened 2023-10-22
Solution py-images-and-custom-auth-serializer

Mazur-Mariia-Python:


cinema/models.py
```diff
@@ -122,9 +139,8 @@ def save(
         )
 
     def __str__(self):
-        return (
-            f"{str(self.movie_session)} (row: {self.row}, seat: {self.seat})"
-        )
+        return f"{str(self.movie_session)} (row: {self.row}," \
+               f" seat: {self.seat})"
```
EdAlekseiev: Use `()` instead of `\` for splitting line

cinema/views.py
```diff
@@ -101,8 +112,30 @@ def get_serializer_class(self):
...
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
EdAlekseiev: You can call `is_valid` with argument `raise_exception=True`. It will automatically return a response with status 400 and list of errors

cinema/views.py
```diff
@@ -101,8 +112,30 @@ def get_serializer_class(self):
...
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST,
+        )
```
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

cinema/views.py
Mazur-Mariia-Python: Thank you for providing the detailed information.

cinema/models.py
```diff
@@ -35,12 +43,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+    return os.path.join("uploads/movies/", filename)
```
ArsenPidhoretskyi, line 49:
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (approved):
Good job!

PR #355, opened 2023-10-25
Solution

vkleshko:


cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def get_path_to_image(instance, file_name):
+    _, ext = os.path.splitext(file_name)
+    file_name = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+    return os.path.join("upload/movies/", file_name)
```
```suggestion
    return os.path.join("upload", "movies", file_name)
```

ArsenPidhoretskyi (changes requested):
Small fix required, because need to make it crossplatform

vasylhnatiuk (approved):

PR #358, opened 2023-11-06
Develop

vlodek69:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Instead of if statement use serializer.is_valid(raise_exception=True)

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

vasylhnatiuk (approved):

PR #361, opened 2023-11-14
Solution

nickkozlov90:


cinema/views.py
```diff
@@ -101,8 +103,22 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, *args, **kwargs):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
NikOneZ1: You can use raise_exception param instead of the if statement
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+
+    return os.path.join("uploads/movies/", filename)
```
NikOneZ1: Do this code crossplatform
```suggestion
    return os.path.join("uploads", "movies", filename)
```

NikOneZ1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #369, opened 2023-11-25
Add an option to upload images, add auth token serializer

KolBohdan:


cinema/views.py
```diff
@@ -94,13 +97,32 @@ def get_queryset(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use raise_exception=True parameter instead of if statement

ArsenPidhoretskyi (changes requested):

cinema/models.py
```diff
@@ -1,6 +1,10 @@
+import os.path
```
Dimosphen1: It is common to import `os` and later call methods from it rather than importing `os.path`.

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+
+    return os.path.join("uploads/movies/", filename)
```
Dimosphen1: Join `uploads` and `movies` with the use of `join` as well, they are the parts of the directory that should not depend on the underlying os.

Dimosphen1 (changes requested):
Several minor changes were requested.

viktoria-rybenchuk (approved):
Well done!

PR #375, opened 2023-12-04
Solution

prochigor:


YuliiaSaliuta (approved):

user/serializers.py
```diff
@@ -10,15 +12,41 @@ class Meta:
...
-
-        return user
+        with transaction.atomic():
+            password = validated_data.pop("password", None)
+            user = super().update(instance, validated_data)
+            if password:
+                user.set_password(password)
+                user.save()
+
+            return user
```
NEkropo1, line 25: great!

user/views.py
```diff
@@ -1,24 +1,55 @@
...
+                    required=True,
+                    location="form",
+                    schema=coreschema.String(
+                        title="Password",
+                        description="Valid password for authentication",
+                    ),
+                ),
+            ],
+            encoding="application/json",
+        )
```
NEkropo1: very strange decision, if you want to provide customized schemas better to move them to settings or other respective modules.

NEkropo1 (changes requested):

user/views.py
prochigor: Yes, I think that these are unnecessary actions in general, I can just use serializer_class in base CreateTokenView

Y-Havryliv (approved):

PR #380, opened 2023-12-05
Solution

D1st3f:


cinema/views.py
```diff
@@ -73,6 +75,17 @@ def _params_to_ints(qs):
         """Converts a list of string IDs to a list of integers"""
         return [int(str_id) for str_id in qs.split(",")]
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk: int = None):
+        movie = self.get_object()
+        serializer = MovieImageSerializer(movie, data=request.data)
+
```
Familenko: Why in this case you make another code in ```def get_serializer_class(self):```?
```
if self.action == "upload_image":
    return MovieImageSerializer
```

cinema/serializers.py
```diff
@@ -33,7 +33,14 @@ class Meta:
...
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = ("id",
+                  "title",
+                  "description",
+                  "duration",
+                  "genres",
+                  "actors",
+                  "image")
+        read_only_fields = ("id", "image")
```
Familenko, line 43: In MovieViewSet you use MovieSerializer in case if it is not list or retrive. Is modification of field ```image``` is possible?

Familenko (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_custom_path(instance, filename):
```
NEkropo1: Since you have hardcoded `"uploads/movies/"` it's not just `create_custom_path` it's more like `create_movie_image_path`

media/uploads/movies/the-departed-3ef0f97e-cb69-4acf-bf23-929c7c8fed6e.png
NEkropo1, line 1: remove redundant files

cinema/serializers.py
NEkropo1: You could just make them read_only...

NEkropo1 (changes requested):

Y-Havryliv (approved):

PR #384, opened 2023-12-05
Develop

Familenko:


SevKrok (approved):
Nice 👍

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImgSerializer,
```
NEkropo1: fix imports

cinema/serializers.py
```diff
@@ -43,15 +52,52 @@ class MovieListSerializer(MovieSerializer):
...
         many=True, read_only=True, slug_field="full_name"
     )
+    image = serializers.SerializerMethodField()
+
+    def get_image(self, obj):
+        request = self.context.get("request")
+        url = request.build_absolute_uri("/")[:-1]
+        if obj.image:
+            return url + path.join(settings.MEDIA_URL, obj.image.name)
+        return None
```
NEkropo1: You already defined your link in `img_directory_path` so model's `ImageField()` and `upload_to` already handling this part, no need of redundant functions.

NEkropo1 (changes requested):
Fix imports, remove redundant functions and redundant files.

cinema/serializers.py
Familenko: It is too simple use ImageField(), I try to make more custom. I will change and save this code for myself only))

Y-Havryliv: serializer should serialize/deserialize data, moving this logic to serializer is not the best decision

Y-Havryliv (approved):

PR #386, opened 2023-12-05
solution

Naz-iv:


DenPrislipskyi (approved):

cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def create_custom_path(instance, filename: str):
```
NEkropo1: since you hardcoded "uploads/movies" its kinda not `create_custom_path` more like `create_path_for_movie` etc.

cinema/views.py
```diff
@@ -101,8 +103,27 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        detail=True,
+        methods=["POST"],
+        url_path="upload-image",
+        permission_classes=[IsAuthenticated]
```
NEkropo1, line 115: Good logic. Can't get why others think that anyone can upload images)

NEkropo1 (changes requested):

NEkropo1 (approved):
Well done!

PR #390, opened 2023-12-05
DRF

DenPrislipskyi:


katyakoziar (approved):
Good job!

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_mage_file_path(instance, filename):
```
Y-Havryliv: I guess there is a typo

user/models.py
```diff
@@ -1,5 +1,48 @@
 from django.contrib.auth.models import AbstractUser
```
Y-Havryliv:
```suggestion
from django.contrib.auth.models import AbstractUser, BaseUserManager
```
and you can delete import below

Y-Havryliv (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
...
+    filename_without_ext, extension = os.path.splitext(filename)
```
Y-Havryliv: I forgot to write when you don't need a variable you can use `_`

Y-Havryliv (approved):

PR #392, opened 2023-12-06
Best custom AuthTokenSerializer ever

Sel-Fisher:


cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def bus_image_file_path(instance, filename):
```
roffi37: i know that we all copy this function from tutorial, but better to rename it, we dont have any buses there)

roffi37 (approved):

cinema/models.py
NEkropo1: yup, rename it.

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
NEkropo1: fix imports

user/serializers.py
```diff
@@ -22,3 +23,41 @@ def update(self, instance, validated_data):
...
+        email = attrs.get("email")
+        password = attrs.get("password")
+
+        if email and password:
+            user = authenticate(request=self.context.get("request"),
+                                email=email, password=password)
+
+            # The authenticate call simply returns None for is_active=False
+            # users. (Assuming the default ModelBackend authentication
+            # backend.)
```
NEkropo1: good comment but redundant.

NEkropo1 (changes requested):

LiudmylaKulinchenko (changes requested):
All checks have failed 😦

PR #402, opened 2023-12-17
Solution

laskarj:


cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance: models.Model, filename: str) -> str:
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+    return os.path.join("uploads/movie_images/", filename)
```
```suggestion
    return os.path.join("uploads", "movie_images", filename)
```

ArsenPidhoretskyi (changes requested):
Fix small comment

laskarj:
I have a question: 
Is there a difference in how to type dict parameters? OrderedDict or just dict, because dict is already sorted by default in Python since version 3.7?
Or does it make no difference in this case?

ArsenPidhoretskyi (approved):
Well done!

PR #404, opened 2023-12-20
Solution

dzhurarostislav:


cinema/models.py
```diff
@@ -1,6 +1,10 @@
+import os.path
```
Dimosphen1, line 1: It is common to import just `os` without further specifications.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #406, opened 2023-12-24
solution

Artemh59:


cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def image_path_create_function(instance, filename):
+    _, extension = os.path.split(filename)
+
+    image_name = f"{slugify(instance.title)}-{uuid.uuid4()}.{extension}"
```
ArsenPidhoretskyi: Is dot really needed?

cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
...
+
+    return os.path.join("uploads/movies/", image_name)
```
```suggestion
    return os.path.join("uploads", "movies", image_name)
```

cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use raise_exception=True instead of if-else statement

ArsenPidhoretskyi (changes requested):

cinema/serializers.py
```diff
@@ -33,7 +33,16 @@ class Meta:
...
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = (
+            "id",
+            "title",
+            "description",
+            "duration",
+            "genres",
+            "actors",
+            "image"
```
viktoria-rybenchuk: add a trailing coma

viktoria-rybenchuk: Fix in all places

viktoria-rybenchuk (changes requested):
Remove all medias files from your repository

cinema/views.py
Dimosphen1: This comment is not encountered, use `raise_exception=True` in the serializer instead of if-else statement to not duplicate the built-in logic.

Dimosphen1 (changes requested):
One minor change was requested.

viktoria-rybenchuk (approved):
Good job!

PR #407, opened 2023-12-28
Solution

Anastasia-Su:


cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser]
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```
vasylhnatiuk: use serializer.is_valid(raise_exception=True):

vasylhnatiuk (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,23 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+
+    return os.path.join("uploads/movies", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

ArsenPidhoretskyi (approved):
Good job

PR #408, opened 2023-12-29
Solution

callogan:


cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, file_name):
+    _, extension = os.path.splitext(file_name)
+    file_name = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+    return os.path.join("uploads/movies/", file_name)
```
```suggestion
    return os.path.join("uploads", "movies", file_name)
```

cinema/views.py
```diff
@@ -101,8 +105,20 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload_image")
+    def upload_image(self, request: Request, pk: int = None) -> Response:
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```
ArsenPidhoretskyi: Use serializer.is_valid(raise_exception=True) instead of if-else statement

cinema/views.py
```diff
@@ -101,8 +105,20 @@ def get_serializer_class(self):
         if self.action == "retrieve":
             return MovieDetailSerializer
 
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload_image")
```
ArsenPidhoretskyi:
```suggestion
    @action(methods=["POST"], detail=True)
```
url_path is redundant because function named the same

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Looks nice to me!

PR #413, opened 2024-01-10
Solution-py-images-and-custom-auth-serializer

Nadiia-developer:


cinema/views.py
```diff
@@ -101,8 +104,27 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
```
pavlejviki: Use raise_exception=True instead of if  statement

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def uploads_movies(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+
+    return os.path.join("uploads/images/", filename)
```
pavlejviki:
```suggestion
    return os.path.join("uploads", "images", filename)
```

pavlejviki (changes requested):
Remove all media files from your repository

Polyakiv-Andrey (approved):
Good job!

PR #421, opened 2024-01-25
Add ImageField image to the Movie model, Replace username field with …

DenysRaboshchuk:
…email field in User model, Write custom AuthTokenSerializer


YukihiroSM (approved):
Looks nice to me!

PR #429, opened 2024-02-15
Solution

nschcur:


cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, ext = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}--{uuid.uuid4()}.{ext}"
```
ArsenPidhoretskyi: Is really dot needed before ext

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
...
+    return os.path.join("uploads/images", filename)
```
```suggestion
    return os.path.join("uploads", "images", filename)
```

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Great!

PR #431, opened 2024-02-21
Solution

Zonda001:


cinema/models.py
```diff
@@ -35,7 +40,23 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{extension}"
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "movies", filename)
```

cinema/views.py
```diff
@@ -78,31 +78,35 @@ def get_queryset(self):
...
-
+        if self.action == "upload_image":
+            return MovieImageSerializer
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk: int):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #432, opened 2024-02-23
Develop

viktor-codes:


EdAlekseiev (changes requested):
Delete `media` folder from repository and add it to .gitignore

EdAlekseiev (approved):
Good job

PR #437, opened 2024-02-27
Solution

kravetb:


EdAlekseiev (changes requested):
Fix tests

kravetb:
I have problems with my test, because on pycharm my tests were done byt on git only two test were done

EdAlekseiev (approved):
Good job

PR #446, opened 2024-03-13
Solution

DanSheremeta:
1. Add image field to Movie model
2. Implemented `upload-image` end-point
3. Replaced username field with email field in User model
4. Implemented AuthTokenSerializer for email field


EdAlekseiev (approved):
Good job

PR #450, opened 2024-03-17
Develop

MaxymChyncha:


ArturPoltser (approved):

cinema/serializers.py
```diff
@@ -44,17 +48,47 @@ class MovieListSerializer(MovieSerializer):
...
+            "description",
+            "duration",
+            "genres",
+            "actors",
+            "image"
+        )
+        read_only_fields = ("id", "image")
+
 
 class MovieDetailSerializer(MovieSerializer):
```
Serhii-Leonenko: Do you use inheritance here and in the previous class?

Serhii-Leonenko (changes requested):

cinema/serializers.py
Serhii-Leonenko: Still inherited from  MovieSerializer, but Meta is overridden

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #451, opened 2024-03-18
Solution

pnasonov:


cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(movie, filename) -> str:
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(movie.title)}-{uuid.uuid4()}.{extension}"
+
+    return os.path.join("uploads/movies/", filename)
```
```suggestion
    return os.path.join("uploads", "images", filename)
```

cinema/views.py
```diff
@@ -101,8 +104,24 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None) -> Response:
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
```
vsmutok: you can use raise_exception=True

vsmutok (changes requested):

cinema/views.py
pnasonov: so should I use raise_exception or not?

vsmutok: no need use if operator here

vsmutok (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(movie, filename) -> str:
+    _, extension = os.path.splitext(filename)
+    filename = f"{slugify(movie.title)}-{uuid.uuid4()}.{extension}"
```
ArsenPidhoretskyi: Is dot really needed before {extension}? Please check output of filename

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #454, opened 2024-03-20
Solution

barabarinov:


cinema/views.py
```diff
@@ -94,14 +97,30 @@ def get_queryset(self):
...
+    def upload_image(self, request, pk: int) -> Response:
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(
+                serializer.data, status=status.HTTP_200_OK
+            )
+
```
kabust, line 114: Return serializer.errors in response in case if serializer is not valid

kabust (changes requested):
Overall good, just one little fix :)

cinema/serializers.py
```diff
@@ -44,17 +55,46 @@ class MovieListSerializer(MovieSerializer):
...
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = (
+            "id",
+            "title",
+            "image",
+            "description",
+            "duration",
+            "genres",
+            "actors",
+        )
```
Serhii-Leonenko: DRY, add image field to MovieSerializer and remove duplicated Meta classes

cinema/views.py
Serhii-Leonenko: no, drf will handle errors, which are raised

Serhii-Leonenko (changes requested):

MNDonut (approved):

PR #462, opened 2024-03-21
Solution(s) (py-images-and-custom-auth-serializer)

baranovr:


baranovr:
Can you give me some advices what I should to do in this case?

I can pass the tests successfully:
Снимок экрана 2024-03-21 025656

But I can't load data from json file:
(django.core.serializers.base.DeserializationError: Problem installing fixture '...\py-images-and-custom-auth-serializer\cinema_service_db_data.json':)
Снимок экрана 2024-03-21 025017

vasylhnatiuk (changes requested):
if you have some questions pls write them in the chat

PR #469, opened 2024-03-21
Solution

isikach:


oksankakot (approved):
Incredible job

cinema/models.py
```diff
@@ -35,12 +40,25 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def image_movie_path(movie: "Movie", filename: str):
+    filename = (
```
MNDonut, line 44: add annotation to return type

user/views.py
```diff
@@ -1,18 +1,20 @@
...
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.settings import api_settings
 
-from user.serializers import UserSerializer
+from user.serializers import UserSerializer, AuthTokenSerializer
 
 
 class CreateUserView(generics.CreateAPIView):
     serializer_class = UserSerializer
+    permission_classes = ()
```
MNDonut, line 12: Why not authenticated user can create a new user?

MNDonut (approved):

PR #472, opened 2024-03-22
Solution

oksankakot:


cinema/serializers.py
```diff
@@ -51,7 +59,21 @@ class MovieDetailSerializer(MovieSerializer):
...
+        fields = (
+            "id",
+            "title",
+            "description",
+            "duration",
+            "genres",
+            "actors",
+            "image"
+        )
+
```
VladKyianytsia: Do you need class Meta here if it is inherited from MovieSerializer?

VladKyianytsia (approved):

cinema/serializers.py
Serhii-Leonenko: agree

user/admin.py
```diff
@@ -1,5 +1,39 @@
 from django.contrib import admin
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 from .models import User
```
Serhii-Leonenko: fix imports

Serhii-Leonenko (changes requested):

user/admin.py
oksankakot: Fixed

cinema/serializers.py
oksankakot: Removed class Meta from here

Y-Havryliv (changes requested):
there is only 1 commit, you forgot to push a new solution

PR #480, opened 2024-03-26
Solution(s) (py-image-and-custom-auth-serializer)

baranovr:


cinema/models.py
```diff
@@ -35,12 +40,22 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def path_to_media(instance, file_name_hex):
+    default_file_name, ext = os.path.splitext(file_name_hex)
+    uuid = uuid4()
+
+    file_name_hex = f"{slugify(instance.title)}-{uuid}{ext}"
+
+    return os.path.join("uploads/movies/", file_name_hex)
```
Dimosphen1, line 47: Join `uploads` and `movies` directories using `os.path.join` as well

cinema/models.py
```diff
@@ -35,12 +40,22 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def path_to_media(instance, file_name_hex):
+    default_file_name, ext = os.path.splitext(file_name_hex)
+    uuid = uuid4()
```
Dimosphen1: Do not create `uuid` variable, call `uuid4()` in the f-string directly

user/migrations/0003_user_username.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

Dimosphen1 (changes requested):
Several changes were requested

Polyakiv-Andrey (approved):
Good job!

PR #483, opened 2024-04-03
Develop

DaniilHinzburh:


EdAlekseiev (changes requested):
Fix flake8

DaniilHinzburh:
Виникає помилка вже в github, локально всі тести пройшли, бібліотеку встановлював та оновив, не допомагає
ERRORS:
cinema.Movie.image: (fields.E210) Cannot use ImageField because Pillow is not installed.
	HINT: Get Pillow at https://pypi.org/project/Pillow/ or run command "python -m pip install Pillow".

System check identified 1 issue (0 silenced).
Error: Process completed with exit code 1.

cinema/views.py
```diff
@@ -98,11 +101,29 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAuthenticated],
+        url_path="upload-image"
+
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(movie: "Movie", filename: str) -> pathlib.Path:
+    ext = pathlib.Path(filename).suffix
+    filename = f"{slugify(movie.title)}-{uuid.uuid4()}{ext}"
+    return pathlib.Path("upload/movies/") / pathlib.Path(filename)
```
```suggestion
    return pathlib.Path("upload") / "movies" / filename
```
Can you try like this?

ArsenPidhoretskyi (changes requested):

cinema/views.py
ArsenPidhoretskyi: Not fixed

ArsenPidhoretskyi (changes requested):
tests are failed
Update requirements, code will not work for other users

ArsenPidhoretskyi (approved):
Well done!

PR #492, opened 2024-04-15
Solution

IjdmLoverI:


cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(movie: "Movie", filename: str):
+    filename = (f"{slugify(movie.title)}-{uuid.uuid4()}"
+                + pathlib.Path(filename).suffix)
+    return pathlib.Path("uploads/movies/") / pathlib.Path(filename)
```
viktoria-rybenchuk: use  `os.path.join`

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(movie: "Movie", filename: str):
+    filename = (f"{slugify(movie.title)}-{uuid.uuid4()}"
+                + pathlib.Path(filename).suffix)
```
viktoria-rybenchuk: use f-string

viktoria-rybenchuk (changes requested):

EdAlekseiev (approved):

PR #493, opened 2024-04-15
done

VladyslavShepilov:


cinema/views.py
```diff
@@ -98,11 +101,23 @@ def get_serializer_class(self):
...
+        elif self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(detail=True, methods=["post"], url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #495, opened 2024-04-20
solution

eugenevsn:


cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(instance: "Movie", filename: str) -> pathlib.Path:
+    filename = (f"{slugify(instance.title)}-{uuid.uuid4()}"
+                + pathlib.Path(filename).suffix
+                )
```
EdAlekseiev: You should use f string instead of concatenation.
```suggestion
    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{pathlib.Path(filename).suffix}"
```

cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
...
+    return pathlib.Path("upload/movies/") / pathlib.Path(filename)
+
+
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
+    image = models.ImageField(upload_to="movie_image_path", null=True)
```
EdAlekseiev: You should pass callable object to `upload_to`
```suggestion
    image = models.ImageField(upload_to=movie_image_path, null=True)
```

cinema/views.py
```diff
@@ -101,16 +104,36 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None) -> Response:
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev: You should pass `raise_exception=True` to is_valid, it returns `Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)` automatically.
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data)
```

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
v1

vasylhnatiuk (approved):

vasylhnatiuk (approved):

PR #496, opened 2024-04-23
sol

tkachuk2291:


cinema/views.py
```diff
@@ -97,12 +99,27 @@ def get_queryset(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="movie-upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```
vasylhnatiuk: use serializer.is_valid(raise_exception=True)

vasylhnatiuk (changes requested):
you don't need push upload_images, delete pls it

cinema/views.py
tkachuk2291: done,thanks

cinema/views.py
```diff
@@ -97,12 +99,27 @@ def get_queryset(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="movie-upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk: you don't need `if` and response 400, if you use serializer.is_valid(raise_exception=True)

vasylhnatiuk: try to avoid it

vasylhnatiuk (changes requested):

cinema/views.py
tkachuk2291: fixied,thanks!

vasylhnatiuk (approved):

PR #499, opened 2024-05-01
Solution

mwellick:


cinema/views.py
```diff
@@ -97,27 +86,42 @@ def get_queryset(self):
...
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request: Request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
```
EdAlekseiev: You don't need `if` statement if you pass `raise_exception=True`
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #500, opened 2024-05-07
Solution

mish-sk:


cinema/views.py
```diff
@@ -101,8 +103,26 @@ def get_serializer_class(self):
...
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = MovieImageSerializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev: You should pass raise_exception=True. Serializer would return`Response` with status `400` automatically if data was invalid

```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #505, opened 2024-05-13
Solution

loplicat:


.gitignore
```diff
@@ -8,3 +8,4 @@ venv/
 **__pycache__/
 *.pyc
 **db.sqlite3
+media
```
viktoria-rybenchuk: add a blank line

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(instance: "Movie", filename: str) -> pathlib.Path:
+    ext = pathlib.Path(filename).suffix
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}{ext}"
+    return pathlib.Path("upload/movies/") / pathlib.Path(filename)
```
viktoria-rybenchuk: Instead of using the / operator, you can use the os.path.join() function, which is a more platform-independent way of joining paths

cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```
viktoria-rybenchuk: use [`raise_exception=true` ](https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true)parameter in is_valid method

viktoria-rybenchuk (changes requested):

.gitignore
loplicat: Thank you for all the tips! I've implemented changes 🙂

cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
+            serializer.save(raise_exception=True)
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #509, opened 2024-05-18
Solution

vburkalo:


EdAlekseiev (changes requested):
1. Fix tests
2. Add media folder to .gitignore and delete from repository. This folder can be large and contain sensitive information

cinema/views.py
```diff
@@ -94,6 +98,25 @@ def get_queryset(self):
...
+        serializer = self.get_serializer(
+            bus,
+            data=request.data,
+            partial=True
+        )
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev: You should pass raise_exception=True. Serializer would returnResponse with status 400 automatically if data was invalid
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #523, opened 2024-05-30
solution

CharmNiovi:


cinema/serializers.py
```diff
@@ -69,6 +81,7 @@ class MovieSessionListSerializer(MovieSessionSerializer):
         source="cinema_hall.capacity", read_only=True
     )
     tickets_available = serializers.IntegerField(read_only=True)
+    movie_image = ImageUploadSerializer(source="movie.image")
```
MNDonut: it's better to use field, not serializer

MNDonut (changes requested):

cinema/views.py
```diff
@@ -101,8 +105,25 @@ def get_serializer_class(self):
...
+        url_path="upload-image"
+    )
+    def upload_image(self, request, pk):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status.HTTP_400_BAD_REQUEST)
```
YuriiKindrat: u could use serializer.is_valid(raise_exception=True) to simplify ur code here

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #526, opened 2024-05-31
Solution 1

semermaks:


cinema/serializers.py
```diff
@@ -36,13 +36,32 @@ class Meta:
...
+
+
 class MovieListSerializer(MovieSerializer):
     genres = serializers.SlugRelatedField(
         many=True, read_only=True, slug_field="name"
     )
     actors = serializers.SlugRelatedField(
         many=True, read_only=True, slug_field="full_name"
     )
+    image = MovieImageSerializer
```
MNDonut: parentheses?
Also no need to use this serializer because it returns movie id(that already is present in this serializer) and image can be a field

MNDonut (changes requested):

MNDonut (approved):

PR #531, opened 2024-06-01
Solution

Chub-andrew:


cinema/views.py
```diff
@@ -101,8 +110,25 @@ def get_serializer_class(self):
...
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
YuriiKindrat: u could use serializer.is_valid(raise_exception=True) to simplify ur code here

YuriiKindrat (changes requested):
all tests should pass

Chub-andrew:
> all tests should pass
githab tests fell due to incorrect import. It seems like everything is fine now.

YuriiKindrat (approved):
GJ

PR #536, opened 2024-06-02
Solution#1

VitalyBashkiser:


cinema/views.py
```diff
@@ -101,8 +105,28 @@ def get_serializer_class(self):
...
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        """Endpoint for uploading image to specific movie"""
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
YuriiKindrat: u could use serializer.is_valid(raise_exception=True) here, to write less code

YuriiKindrat (changes requested):
remove media from ur PR

cinema/views.py
YuriiKindrat: u didn't fix that

YuriiKindrat (changes requested):

cinema/views.py
VitalyBashkiser: I added raise_Exception=True to is_valid(), maybe this change didn't make it into git

VitalyBashkiser: Or did I misunderstand the error?

YuriiKindrat (approved):
GJ

PR #537, opened 2024-06-03
Solution

ghost:


EdAlekseiev (changes requested):
1. Add media folders to .gitignore and delete it from repository

EdAlekseiev (approved):

PR #558, opened 2024-06-27
Solution

Artem-user-19:


cinema/views.py
```diff
@@ -94,13 +96,31 @@ def get_queryset(self):
...
+        url_name="upload-image"
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OKK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
viktoria-rybenchuk: use raise_exception=True in `is_valid` method - https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true

viktoria-rybenchuk (changes requested):
Add `Pillow` to the requirements

viktoria-rybenchuk (approved):
GJ!

PR #567, opened 2024-07-15
Solution

frontbastard:


cinema/views.py
```diff
@@ -101,8 +95,30 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        bus = self.get_object()
+        serializer = self.get_serializer(bus, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev: You can simplify your code with using is_valid with raise_exception=True
https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

EdAlekseiev (changes requested):

vsmutok (approved):
Looks Good!

PR #581, opened 2024-07-31
Solution_v1

BileichukIvan:


user/serializers.py
```diff
@@ -22,3 +31,38 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```
MNDonut, line 37: you can use EmailField here

MNDonut (approved):
One improvement comment left, read please ;)

PR #582, opened 2024-07-31
Solution

VShveda:


cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
...
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
+    image = models.ImageField(null=True, upload_to=create_custom_path)
+
+    def get_image_url(self):
+        return self.image.url
```
MNDonut: do you use it somewhere?

cinema_service/settings.py
```diff
@@ -125,13 +125,15 @@
...
-USE_TZ = False
+USE_TZ = True
 
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/4.0/howto/static-files/
 
 STATIC_URL = "static/"
+MEDIA_URL = "/media/"
+MEDIA_ROOT = BASE_DIR / "media"
```
MNDonut, line 137: you should ignore media files, so let's remove them(because it's your local fiels and we don't need them here) + ignore in the future commits

user/serializers.py
```diff
@@ -22,3 +31,38 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

MNDonut (changes requested):

Arsen-hrynevych (approved):
Thanks

PR #583, opened 2024-07-31
solution

necobra:


cinema/serializers.py
```diff
@@ -12,6 +12,12 @@
 )
 
 
+class MovieImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ["id", "image"]
```
MNDonut: let's be consistent and use either () or [] everywhere

cinema/serializers.py
```diff
@@ -33,7 +39,9 @@ class Meta:
 class MovieSerializer(serializers.ModelSerializer):
     class Meta:
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = ("id", "title", "description", "duration",
+                  "genres", "actors", "image")
```
MNDonut: we don't add fields in this way. the best practice is to do like you do with imports(1 import - 1line)

cinema/views.py
```diff
@@ -101,8 +104,25 @@ def get_serializer_class(self):
...
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
MNDonut: you can pass parameter raise_exception=True in is_valid and remove if statement because it'll raise error if something goes wrong

user/serializers.py
```diff
@@ -22,3 +25,41 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

user/serializers.py
```diff
@@ -22,3 +25,41 @@ def update(self, instance, validated_data):
...
+        read_only=True
+    )
+
+    def validate(self, attrs):
+        email = attrs.get("email")
+        password = attrs.get("password")
+
+        if email and password:
+            user = authenticate(request=self.context.get("request"),
+                                email=email, password=password)
```
MNDonut: 1 parameter - 1 line

cinema_service/settings.py
```diff
@@ -125,14 +125,18 @@
...
 
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/4.0/howto/static-files/
 
 STATIC_URL = "static/"
 
+MEDIA_ROOT = BASE_DIR / "media"
+
+MEDIA_URL = "/media/"
```
MNDonut, line 138: don't push media to the repo because it's your local data and it's not needed here

MNDonut (changes requested):

cinema/serializers.py
necobra: but the line is too long

cinema/views.py
necobra: explicit better than implicit?

cinema/serializers.py
```diff
@@ -51,7 +66,8 @@ class MovieDetailSerializer(MovieSerializer):
 
     class Meta:
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = ("id", "title", "description", "duration",
+                  "genres", "actors", "image")
```
Arsen-hrynevych, line 70: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

user/models.py
```diff
@@ -1,5 +1,55 @@
...
+from django.contrib.auth.models import (
+    UserManager as DjangoUserManager,
+    AbstractUser
+)
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+# default manager is django.contrib.auth.models.UserManager
+# you can check code to understand what we change
```
Arsen-hrynevych, line 11: This comment shouldn't be there

user/serializers.py
```diff
@@ -1,13 +1,16 @@
...
 
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = get_user_model()
         fields = ("id", "email", "password", "is_staff")
         read_only_fields = ("is_staff",)
-        extra_kwargs = {"password": {"write_only": True, "min_length": 5}}
+        extra_kwargs = {"password": {"write_only": True,
+                                     "min_length": 5,
+                                     "style": {"input_type": "password"}}}
```
Arsen-hrynevych, line 13: Please format code

user/serializers.py
```diff
@@ -22,3 +25,44 @@ def update(self, instance, validated_data):
...
+        if email and password:
+            user = authenticate(
+                request=self.context.get("request"),
+                email=email,
+                password=password
+            )
+
+            # The authenticate call simply returns None for is_active=False
+            # users. (Assuming the default ModelBackend authentication
+            # backend.)
```
Arsen-hrynevych, line 59: This comment shouldn't be there

Arsen-hrynevych (approved):
Overall good

PR #584, opened 2024-07-31
add solution

jpdreamthug:


user/serializers.py
```diff
@@ -22,3 +33,38 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"), write_only=True)
```
MNDonut, line 39: you can use EmailField here

MNDonut (approved):
One improvement comment left, read it please ;)

PR #588, opened 2024-08-01
Solution v.1

SemenetsSergii:


SemenetsSergii:
image
have this problem (after search info at Google add pytest.ini. after play with version django & djangorestframework)

cinema/views.py
```diff
@@ -101,8 +115,27 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

DevSpaciX (changes requested):
Delete pls all images from PR ) And folder media/upload/movies too ))

user/serializers.py
```diff
@@ -22,3 +31,39 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

MNDonut (approved):

PR #589, opened 2024-08-01
Develop

BohdanKuzik:


cinema/views.py
```diff
@@ -97,12 +100,26 @@ def get_queryset(self):
...
+        detail=True,
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

DevSpaciX (changes requested):
Remove from PR media/uploads/movies folder

cinema/serializers.py
```diff
@@ -33,7 +39,9 @@ class Meta:
 class MovieSerializer(serializers.ModelSerializer):
     class Meta:
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = ("id", "title", "description", "duration",
+                  "genres", "actors", "image")
```
MNDonut: the best practice is to move each field on a separate line like you do with imports when it's longer than 79 characters ;)

cinema/views.py
MNDonut: still not fixed

user/serializers.py
```diff
@@ -22,3 +30,47 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

MNDonut (changes requested):

user/admin.py
```diff
@@ -1,5 +1,26 @@
 from django.contrib import admin
-from django.contrib.auth.admin import UserAdmin
-from .models import User
+from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
+from django.utils.translation import gettext as _
+from user.models import User
```
Arsen-hrynevych, line 4: Please add a blank line between them.

Arsen-hrynevych (approved):
Thanks

PR #591, opened 2024-08-01
solution

Andrii1000:


cinema/views.py
```diff
@@ -98,11 +101,22 @@ def get_serializer_class(self):
...
         return MovieSerializer
 
+    @action(detail=True, methods=["post"], url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

DevSpaciX (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,24 @@ def full_name(self):
...
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
     duration = models.IntegerField()
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
+    image = models.ImageField(upload_to=create_custom_path)
+
+    def get_image_url(self):
+        return self.image.url
```
MNDonut: do you use it somewhere?

user/serializers.py
```diff
@@ -22,3 +30,41 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

MNDonut (changes requested):

cinema/models.py
Andrii1000: i think that get_image_url function is useful to provide a convenient way to access the URL of the movie's image. But now I understand that don`t need that

Arsen-hrynevych (approved):
Thanks

PR #592, opened 2024-08-01
Solution

SrMusienko:


cinema/views.py
```diff
@@ -101,8 +114,25 @@ def get_serializer_class(self):
...
 
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        methods=["post"],
+        detail=True,
+        permission_classes=[],
```
DevSpaciX: if you have no permission_classes , you can just delete it

cinema/views.py
```diff
@@ -101,8 +114,25 @@ def get_serializer_class(self):
...
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```

DevSpaciX (changes requested):

user/models.py
```diff
@@ -1,5 +1,46 @@
...
+            raise ValueError("Superuser must have is_staff=True.")
+
+        if extra_fields.get("is_superuser") is not True:
+            raise ValueError("Superuser must have is_superuser=True.")
+
+        return self._create_user(email, password, **extra_fields)
 
 
 class User(AbstractUser):
     pass
```
MNDonut: remove this pass

user/serializers.py
```diff
@@ -20,5 +27,36 @@ def update(self, instance, validated_data):
         if password:
             user.set_password(password)
             user.save()
-
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label="Email", write_only=True)
```

MNDonut (changes requested):

Arsen-hrynevych (approved):
Thanks

PR #593, opened 2024-08-01
Solution

YevaYe:


cinema/views.py
```diff
@@ -101,8 +103,27 @@ def get_serializer_class(self):
...
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
DevSpaciX: no need to use if here
better use serializer.is_valid(raise_exception=True)

DevSpaciX (changes requested):

cinema/views.py
```diff
@@ -21,7 +23,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
 )
```
MNDonut: we don't add multiple imports in 1 line when use ()

user/serializers.py
```diff
@@ -22,3 +30,36 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"), write_only=True)
```

MNDonut (changes requested):

MNDonut (approved):
Awesome ;)

PR #597, opened 2024-08-01
Solution

SuskyiVolodymyr:


cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request: Request, pk: int = None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
DevSpaciX, line 119: no need to use if here
better use `serializer.is_valid(raise_exception=True)`

DevSpaciX (approved):

PR #598, opened 2024-08-01
'Solution'

Irina17191:


cinema/views.py
```diff
@@ -101,8 +103,27 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

DevSpaciX (changes requested):

cinema/views.py
```diff
@@ -101,8 +103,25 @@ def get_serializer_class(self):
...
+        url_path="upload-image",
+        permission_classes=[IsAdminUser],
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
MNDonut: no need to use is_valid in if statement
it's enough to run it like this:
```
serializer.is_valid(..)
serializer.save()
...
```

user/serializers.py
```diff
@@ -22,3 +23,33 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"))
```

MNDonut (changes requested):

DevSpaciX (approved):
Nice

PR #599, opened 2024-08-02
solution

NikitaNikolaitsev:


cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request: Request, pk: int = None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

DevSpaciX (changes requested):
Fix all flake8 errors , and tests too

cinema/views.py
```diff
@@ -21,7 +24,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
MNDonut: we don't add 2+ imports on the same line when use ()

cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request: Request, pk: int = None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
MNDonut: no need to pass is_valid in if statement. you can just call this method separately and remove if statement at all because if there is an error, it'll be raised and serializer.save() will not be reachable

user/serializers.py
```diff
@@ -22,3 +24,38 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```
MNDonut, line 30: you can use EmailField here

MNDonut (changes requested):

Arsen-hrynevych (approved):
Thanks

PR #600, opened 2024-08-02
Develop

vladyslavchichkan:


cinema/views.py
```diff
@@ -21,7 +24,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
MNDonut: we don't add 2+ imports on the same line when use ()

cinema/views.py
```diff
@@ -97,12 +100,27 @@ def get_queryset(self):
...
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        url_path="upload-image"
+    )
+    def upload_image(self, request, pk=None):
+        movie = get_object_or_404(Movie, pk=pk)
+        serializer = self.get_serializer(movie, data=request.data)
+        print(request.data)
```
MNDonut: you can remove this print ;)

cinema/views.py
```diff
@@ -97,12 +100,27 @@ def get_queryset(self):
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
MNDonut: you can simplify it calling serializer.is_valid(raise_exception=True) separately, thus there won't be a need to add if statement

user/serializers.py
```diff
@@ -22,3 +31,41 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```

MNDonut (changes requested):

Arsen-hrynevych (approved):
Thanks

PR #601, opened 2024-08-04
Solution

lilarin:


cinema/models.py
```diff
@@ -42,6 +46,15 @@ class Movie(models.Model):
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
 
+    def movie_image_path(self: "Movie", filename: str) -> pathlib.Path:
+        filename = pathlib.Path(
+            f"{slugify(self.title)}-{uuid.uuid4()}"
+            + pathlib.Path(filename).suffix
+        )
+        return pathlib.Path("movies/") / filename
```
MNDonut: The best practice is to move it out of this class so it's reusable by other models

user/serializers.py
```diff
@@ -7,18 +10,57 @@ class Meta:
...
         user = super().update(instance, validated_data)
         if password:
             user.set_password(password)
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label="Email address", write_only=True)
```

MNDonut (changes requested):
Two small changes ;)

cinema/models.py
lilarin: Got it

DevSpaciX (approved):
GJ !

PR #604, opened 2024-08-07
sol1

IvanVahanov:


cinema/views.py
```diff
@@ -103,6 +107,31 @@ def get_serializer_class(self):
...
+            return Response(
+                {"detail": "Uploaded file is not a valid image."},
+                status=status.HTTP_400_BAD_REQUEST
+            )
+        movie.image = image
+        movie.save()
+        return Response({
+            "status": "image uploaded",
+            "image": movie.image.url
+        }, status=status.HTTP_200_OK)
```
DevSpaciX, line 133: You can provide here just url_path="upload-image", and then in get_serializer_class use MovieImageSerializer if self.action will be upload-image . And in this func just check if serializer is valid will be ok )

DevSpaciX (changes requested):
Pls do not push test images in PR

PR #623, opened 2024-09-13
Implement solution

dimak20:


cinema/views.py
```diff
@@ -101,8 +103,27 @@ def get_serializer_class(self):
...
+            "POST",
+        ],
+        detail=True,
+        permission_classes=[IsAdminOrIfAuthenticatedReadOnly],
+        url_path="movie-upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, request.data)
+        if serializer.is_valid():
```
oliasymak24: raise_exception=True will be better)

oliasymak24 (changes requested):

cinema/views.py
dimak20: Yea, firstly did exactly throughh this kwarg, but couldn't find out how to send a json-response like error

cinema/views.py
```diff
@@ -101,8 +103,26 @@ def get_serializer_class(self):
...
+
         return MovieSerializer
 
+    @action(
+        methods=[
+            "POST",
+        ],
+        detail=True,
+        permission_classes=[IsAdminOrIfAuthenticatedReadOnly],
+        url_path="movie-upload-image",
```
Oleksl888, line 117: > Upload images  should be only available with /upload-image/ endpoint.

user/models.py
```diff
@@ -1,5 +1,50 @@
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import (
+    AbstractUser,
+    UserManager as DjangoUserManger
+)
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class UserManager(DjangoUserManger):
```
Oleksl888, line 9: Would be better structure if you put manager in separate file

Oleksl888 (approved):
Approved. Fix the small issues please

PR #626, opened 2024-09-24
Implemented images managing for movie and custom user without username

MasakDirt:


cinema/serializers.py
```diff
@@ -33,7 +33,16 @@ class Meta:
...
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = (
+            "id",
+            "title",
+            "description",
+            "duration",
+            "genres",
+            "actors",
+            "image"
```

cinema/serializers.py
```diff
@@ -51,7 +60,22 @@ class MovieDetailSerializer(MovieSerializer):
...
+        )
+        read_only_fields = ("id", "image")
+
+
+class MovieUploadImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ("image",)
```
Oleksl888: "id" is also expected

cinema/views.py
```diff
@@ -21,7 +24,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieUploadImageSerializer,
```
Oleksl888: each import must be on new line

requirements.txt
Oleksl888, line 1: why is this binary?

user/models.py
```diff
@@ -1,5 +1,43 @@
+from django.contrib.auth.base_user import BaseUserManager
 from django.contrib.auth.models import AbstractUser
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class UserManager(BaseUserManager):
```

Oleksl888 (changes requested):

cinema/models.py
```diff
@@ -35,12 +39,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def get_movie_image_path(movie: "Movie", filename: str) -> pathlib.Path:
+    extension = pathlib.Path(filename).suffix
+    image_name = f"{slugify(movie.title)}-{uuid.uuid4()}{extension}"
+    return pathlib.Path("uploads/movies") / pathlib.Path(image_name)
+
```
Nattalli, line 47: suggest you use os.path.join to avoid errors related to the use of different OS

user/serializers.py
```diff
@@ -22,3 +23,37 @@ def update(self, instance, validated_data):
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(
```
Nattalli: you can use `serializers.EmailField`

Nattalli (changes requested):

Oleksl888 (approved):
Well done!

PR #630, opened 2024-10-01
solution

BastovOleksandr:


cinema/serializers.py
```diff
@@ -30,10 +30,26 @@ class Meta:
...
 
 
+class MovieImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ("id", "image")
+
+
 class MovieSerializer(serializers.ModelSerializer):
+    image = serializers.CharField(read_only=True)
```
Oleksl888: you don't need to include image here - this serializer is used for creating new movie without image  as per task requirements

cinema/serializers.py
```diff
@@ -111,10 +141,18 @@ class MovieSessionDetailSerializer(MovieSessionSerializer):
     taken_places = TicketSeatsSerializer(
         source="tickets", many=True, read_only=True
     )
+    movie_image = MovieImageSerializer(many=False, read_only=True)
```
Oleksl888: as per task requirements you need to display image here:
So you already have an image on MovieListSerializer but hen you add an extra one
> Movie session: on list page with movie_image key; on detail page inside movie -> image.

user/models.py
```diff
@@ -1,5 +1,45 @@
+from django.contrib.auth.base_user import BaseUserManager
 from django.contrib.auth.models import AbstractUser
+from django.utils.translation import gettext as _
+from django.db import models
+
+
+class UserManager(BaseUserManager):
```
Oleksl888: Would be better structure if you put Manager in a separate file

Oleksl888 (changes requested):

cinema/views.py
```diff
@@ -101,8 +104,24 @@ def get_serializer_class(self):
...
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer_class()(movie, data=request.data)
```
Oleksl888, line 120: here you can call `get_serializer(` to make it simpler

Oleksl888 (approved):
Good job!

PR #636, opened 2024-10-03
Solution

OleksiukStepan:


cinema/serializers.py
```diff
@@ -44,14 +57,16 @@ class MovieListSerializer(MovieSerializer):
...
 
 class MovieDetailSerializer(MovieSerializer):
     genres = GenreSerializer(many=True, read_only=True)
     actors = ActorSerializer(many=True, read_only=True)
 
-    class Meta:
-        model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+    class Meta(MovieSerializer.Meta):
+        fields = MovieSerializer.Meta.fields + ("image",)
```
BastovOleksandr: I'm not sure, but if we use this in both: MovieDetail and MovieList serializer classes, maybe it is worth moving the "image" field to MovieSerializer fields? to the row 42
This is an assumption, to change it or not is at your discretion)

cinema/views.py
```diff
@@ -101,8 +105,24 @@ def get_serializer_class(self):
...
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminOrIfAuthenticatedReadOnly],
+        url_path="upload-image",
```
BastovOleksandr: Maybe the IsAdminUser permission class fits better here?
Authenticated users cant use the POST method anyway

user/models.py
```diff
@@ -1,5 +1,52 @@
...
+    AbstractUser,
+    UserManager as DjangoUserManager
+)
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class UserManager(DjangoUserManager):
+    """Define a model manager for User model with no username field."""
+
```
BastovOleksandr: I don't even know if there is any point in deleting this, but I'll make a note)
I mean docstrings

BastovOleksandr (approved):

user/models.py
OleksiukStepan: I believe it's acceptable to leave this comment as we include additional information about why we implemented this class

cinema/serializers.py
```diff
@@ -31,9 +31,25 @@ class Meta:
 
 
 class MovieSerializer(serializers.ModelSerializer):
+    image = serializers.ImageField(read_only=True)
```

cinema/serializers.py
```diff
@@ -31,9 +31,25 @@ class Meta:
...
+        )
+
+
+class MovieImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ["id", "image"]
 
 
 class MovieListSerializer(MovieSerializer):
```
Oleksl888, line 52: image needs to be on list and detail serializers

user/models.py
```diff
@@ -1,5 +1,52 @@
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import (
+    AbstractUser,
+    UserManager as DjangoUserManager
+)
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class UserManager(DjangoUserManager):
```

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #643, opened 2024-10-05
Solution

pashawarganov:


DSahalatyi (approved):

cinema/serializers.py
```diff
@@ -31,9 +31,20 @@ class Meta:
 
 
 class MovieSerializer(serializers.ModelSerializer):
+    image = serializers.ImageField(read_only=True)
```

cinema/serializers.py
```diff
@@ -31,9 +31,20 @@ class Meta:
...
+        )
+
+
+class MovieImageSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ("id", "image")
 
 
 class MovieListSerializer(MovieSerializer):
```
Oleksl888: image must be included here too

cinema/views.py
```diff
@@ -101,8 +93,26 @@ def get_serializer_class(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[permissions.IsAdminUser],
+        url_path="upload-image"
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
Oleksl888: no need for if because it will never be evaluated as False because of exception

cinema/views.py
```diff
@@ -101,8 +93,26 @@ def get_serializer_class(self):
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
Oleksl888: this will never be reached because of raise_exception

user/models.py
```diff
@@ -1,5 +1,53 @@
-from django.contrib.auth.models import AbstractUser
+from django.contrib.auth.models import (
+    AbstractUser,
+    UserManager as DjangoUserManager
+)
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class UserManager(DjangoUserManager):
```

Oleksl888 (changes requested):

cinema/serializers.py
pashawarganov: do I need included it here, if I override class Meta?

Oleksl888: short answer is yes :+1:

Oleksl888 (approved):
Well done sir!

PR #645, opened 2024-10-05
'Solution'

dryzhenko:


dryzhenko:
if I don't add pillow to requirements.txt, tests don't pass, and when I add it, "install requirements" doesn't pass, tell me how to fix it please

vsmutok (changes requested):
Tests Failed!

Add `pillow` to `requirements.txt`

cinema/views.py
```diff
@@ -101,8 +104,26 @@ def get_serializer_class(self):
...
+        detail=True,
+        permission_classes=[IsAdminUser],
+        url_path="upload-image",
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
vsmutok: Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #649, opened 2024-10-06
first

egor250330:


Bogdan-Rylov (approved):

cinema/models.py
```diff
@@ -35,12 +39,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movies_image_path(movie: "Movie", filename: str) -> pathlib.Path:
+    filename = (
+        f"{slugify(movie.title)}-{uuid.uuid4()}"
+        + pathlib.Path(filename).suffix
+    )
+    return pathlib.Path("upload-image/") / pathlib.Path(filename)
```
Nattalli: suggest you use os.path.join to avoid errors related to the use of different OS

user/models.py
```diff
@@ -1,5 +1,47 @@
...
+        return user
+
+    def create_user(self, email, password=None, **extra_fields):
+        extra_fields.setdefault("is_staff", False)
+        extra_fields.setdefault("is_superuser", False)
+        return self._create_user(email, password, **extra_fields)
+
+    def create_superuser(self, email, password, **extra_fields):
+        extra_fields.setdefault("is_staff", True)
+        extra_fields.setdefault("is_superuser", True)
```
Nattalli: suggest you remove managers to `managers.py`

user/serializers.py
```diff
@@ -1,24 +1,62 @@
...
         user = super().update(instance, validated_data)
         if password:
             user.set_password(password)
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"), write_only=True)
```
Nattalli: you can use `serializers.EmailField`

Nattalli (changes requested):

cinema/models.py
```diff
@@ -35,12 +40,23 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movies_image_path(movie: "Movie", filename: str) -> str:
```
Oleksl888: Better to move this to separate file utils.py or helpers.py

cinema/serializers.py
```diff
@@ -31,9 +31,25 @@ class Meta:
...
         model = Movie
-        fields = ("id", "title", "description", "duration", "genres", "actors")
+        fields = (
+            "id",
+            "title",
+            "description",
+            "duration",
+            "genres",
+            "actors",
+            "image"
```
Oleksl888: As per task requirements:

> Image field should not be available on POST api/cinema/movies/.
> Image url should be shown on:
> Movie: list and detail pages

With your implementation it is being shown on every endpoint. Make sure you show it only on List/Detail

Oleksl888 (changes requested):

cinema/helper.py
```diff
@@ -0,0 +1,17 @@
+import os
+import pathlib
+import uuid
+from typing import Optional, TYPE_CHECKING
+from django.utils.text import slugify
+
+if TYPE_CHECKING:
+    from cinema.models import Movie
```
Oleksl888, line 8: by doing this you create circular dependency -> here you import Movie from models and in models you import movies_image_path from helper

Oleksl888: by doing this you create circular dependency -> here you import Movie from models and in models you import movies_image_path from helper

cinema/serializers.py
Oleksl888: I suggest you just remove image field from here, then add it on MovieListSerializer and MovieDetailSerializer below and so you will not need to make a new MovieCreateSerializer

Oleksl888 (changes requested):

MNDonut (approved):

PR #657, opened 2024-10-07
'Solution'

Poznanskyi:


user/migrations/0001_initial.py
olena4reunited, line 1: Why you delete migration? You need to run makemigrations again

olena4reunited (changes requested):
overall, good job

cinema/models.py
```diff
@@ -35,12 +38,19 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def get_movie_image_path(instance, filename) -> pathlib.Path:
+    filename = (f"slugify {instance.title}-{uuid.uuid4()}"
+                + pathlib.Path(filename).suffix)
+    return pathlib.Path("uploads/movies/") / pathlib.Path(filename)
```
Nattalli: suggest you use os.path.join to avoid errors are related to the use of different OS

cinema/views.py
```diff
@@ -94,13 +97,30 @@ def get_queryset(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
```
Nattalli: it would be better to add new functions after overwritten function (get_serializer_class in your case)

Nattalli: you can use raise_exception=True

media/uploads-movies/sample-movie-1aaa243c-0d93-4966-98e9-d567f3e96992.jpg
Nattalli, line 1: probably you should remove that and add to gitignore

user/serializers.py
```diff
@@ -1,24 +1,62 @@
...
         user = super().update(instance, validated_data)
         if password:
             user.set_password(password)
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"), write_only=True)
```
Nattalli: you can use serializers.EmailField

Nattalli (changes requested):

cinema/models.py
```diff
@@ -35,12 +38,26 @@ def full_name(self):
...
+def get_movie_image_path(instance, filename) -> pathlib.Path:
+    unique_filename = (
+        f"{slugify(instance.title)}-"
+        f"{uuid.uuid4()}"
+        f"{pathlib.Path(filename).suffix}"
+    )
+    return pathlib.Path("uploads/movies") / unique_filename
+
+
+def slugify(value: str) -> str:
```
Oleksl888, line 50: why you don't reuse the built in slugify? it covers more cases

Oleksl888 (approved):
well done sir :handshake:

PR #693, opened 2024-12-09
Solution

VladyslavBon:
Solution


cinema/views.py
```diff
@@ -101,8 +110,27 @@ def get_serializer_class(self):
...
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = MovieImageSerializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
vladyslav-tmf, line 132: Consider simplifying validation by using `raise_exception=True` parameter in serializer.
It will automatically throw a ValidationError with HTTP 400 on invalid data, which makes the code cleaner and avoids explicit validation checks.

vladyslav-tmf (approved):

PR #696, opened 2024-12-11
implement images and custom auth token serializers

Mxbely:


VladyslavBon (approved):

cinema/models.py
```diff
@@ -35,12 +38,23 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_file_path(instance, filename):
+    filename = (
+        f"{slugify(instance.title)}-{uuid.uuid4()}"
+        + pathlib.Path(filename).suffix
+    )
+    return pathlib.Path("upload/movies/") / pathlib.Path(filename)
```
AnyoneClown: You are converting `filename` 2 times here, you don't need to do it in second one(return statment)

cinema/views.py
```diff
@@ -101,8 +105,21 @@ def get_serializer_class(self):
...
+
         return MovieSerializer
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
```
AnyoneClown: Use `serializer.is_valid(raise_exception=True)`

user/views.py
```diff
@@ -4,15 +4,17 @@
 from rest_framework.authtoken.views import ObtainAuthToken
 from rest_framework.settings import api_settings
 
-from user.serializers import UserSerializer
+from user.serializers import UserSerializer, AuthTokenSerializer
 
 
 class CreateUserView(generics.CreateAPIView):
     serializer_class = UserSerializer
+    permission_classes = ()
```
AnyoneClown: Don't add blank attributes

AnyoneClown (changes requested):

Nattalli (approved):

PR #697, opened 2024-12-12
Solution

pohgen:


cinema/models.py
```diff
@@ -35,12 +40,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image(instance: "Movie", filename: str) -> Path:
+    filename = (
+        f"{slugify(instance.title)}-{uuid.uuid4()}"
+        + pathlib.Path(filename).suffix
+    )
+    return pathlib.Path("upload/movies/") / pathlib.Path(filename)
```
Nattalli: so many `pathlib.Path` usage, don't overuse it. You can use `os.path.join` here

cinema/views.py
```diff
@@ -94,13 +97,25 @@ def get_queryset(self):
 
         return queryset.distinct()
 
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
Nattalli: use `serializer.is_valid(raise_exception=True)`

Nattalli (changes requested):

cinema/views.py
```diff
@@ -101,7 +101,7 @@ def get_queryset(self):
     def upload_image(self, request, pk=None):
         movie = self.get_object()
         serializer = self.get_serializer(movie, data=request.data)
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
             serializer.save()
             return Response(serializer.data, status=status.HTTP_200_OK)
```
Nattalli: serializer.is_valid(raise_exception=True)
        serializer.save()

Nattalli (changes requested):

AnyoneClown (approved):
LGTM

PR #706, opened 2024-12-15
images-and-custom-auth-serializer solution

94nj111:


media/upload/movies/sample-movie-05399760-0fdc-4399-ae05-64373c739de6.jpg
Den-k0, line 1: I think, you don't need to add this image and another 17 images to git

Den-k0 (approved):
lgtm

PR #710, opened 2024-12-15
Solution

YaroslavBordovoy:


AndriiDementiev (approved):

cinema/migrations/0002_movie_image.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('cinema', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='movie',
+            name='image',
+            field=models.ImageField(null=True, upload_to=cinema.models.movie_image_path),
```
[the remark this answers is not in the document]
YaroslavBordovoy: This is a migration file, it is created automatically. The logic in the model is implemented correctly

cinema/models.py
```diff
@@ -1,6 +1,16 @@
...
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.conf import settings
+from django.utils.text import slugify
+
+
+def movie_image_path(instance: "Movie", filename: str) -> pathlib.Path:
+    filename = (f"{slugify(instance.title)}-{uuid.uuid4()}"
+                + pathlib.Path(filename).suffix)
+    return pathlib.Path("uploads/movies/") / pathlib.Path(filename)
```
Nattalli: better use `os.path.join` here

user/models.py
```diff
@@ -1,5 +1,53 @@
...
+        return self._create_user(email, password, **extra_fields)
 
 
 class User(AbstractUser):
-    pass
+    username = None
+    email = models.EmailField(_("email address"), unique=True)
+
+    USERNAME_FIELD = "email"
+    REQUIRED_FIELDS = []
```
Nattalli, line 51: Isn't `email` required?

user/views.py
```diff
@@ -4,15 +4,17 @@
 from rest_framework.authtoken.views import ObtainAuthToken
 from rest_framework.settings import api_settings
 
-from user.serializers import UserSerializer
+from user.serializers import UserSerializer, AuthTokenSerializer
 
 
 class CreateUserView(generics.CreateAPIView):
     serializer_class = UserSerializer
+    permission_classes = ()
```
Nattalli: don't leave it empty. Remove or specify something insade

Nattalli (changes requested):
remove static files, add that folder to gitignore

user/models.py
YaroslavBordovoy: If we defined USERNAME_FIELD as email, then there is no need to duplicate it in REQUIRED_FIELDS. In this case, even migration is not possible.

user/views.py
YaroslavBordovoy: Done

cinema/models.py
YaroslavBordovoy: Done

AnyoneClown (approved):
Good job

PR #720, opened 2024-12-16
Solution

Rabb1t-Artem:


media/uploads/movies/sample-movie-042a6922-2a9f-4081-8b64-800d22a5df1a.jpg
Den-k0, line 1: I think, you don't need to add this 3 images to git

Den-k0 (approved):
lgtm

PR #725, opened 2024-12-17
Solution

yulia-pl:


cinema/models.py
```diff
@@ -38,16 +44,25 @@ def full_name(self):
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    duration = models.IntegerField()
+    duration = models.IntegerField(default=90)
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
+    image = models.ImageField(upload_to="uploads/movies/",
+                              null=True, blank=True)
```
dxrrkwm: blank true is enough here

cinema/models.py
```diff
@@ -38,16 +44,25 @@ def full_name(self):
...
     def __str__(self):
         return self.title
 
+    def save(self, *args, **kwargs):
+        if self.image:
+            ext = os.path.splitext(self.image.name)[1]
+            unique_id = uuid.uuid4()
+            self.image.name = f"{slugify(self.title)}-{unique_id}{ext}"
+        super().save(*args, **kwargs)
+
```
dxrrkwm: use more descriptive name for method and use pathlib.Path instead of os

cinema/models.py
```diff
@@ -57,15 +72,22 @@ class MovieSession(models.Model):
     class Meta:
         ordering = ["-show_time"]
 
+    def save(self, *args, **kwargs):
+        if isinstance(self.show_time, str):
+            self.show_time =\
+                datetime.strptime(self.show_time, "%Y-%m-%d %H:%M:%S")
+        if timezone.is_naive(self.show_time):
+            self.show_time = timezone.make_aware(self.show_time)
+        super().save(*args, **kwargs)
```
dxrrkwm, line 80: ? don't need it here, the error with naive timezone is ok in this task

cinema/models.py
```diff
@@ -78,9 +100,8 @@ class Ticket(models.Model):
     movie_session = models.ForeignKey(
         MovieSession, on_delete=models.CASCADE, related_name="tickets"
     )
-    order = models.ForeignKey(
-        Order, on_delete=models.CASCADE, related_name="tickets"
-    )
+    order = models.ForeignKey(Order, on_delete=models.CASCADE,
+                              related_name="tickets")
```
dxrrkwm, line 103: we don't have problem with long line here, why changing it?
same in other files and this one in particular

cinema/views.py
```diff
@@ -103,6 +110,10 @@ def get_serializer_class(self):
 
         return MovieSerializer
 
+    def perform_create(self, serializer):
+        """Ensure the image field is not set during movie creation"""
+        serializer.save(image=None)
```
dxrrkwm, line 115: it's better to write upload image here with action decorator

cinema/views.py
```diff
@@ -173,3 +184,17 @@ def get_serializer_class(self):
...
+class MovieImageUploadView(APIView):
+    parser_classes = [MultiPartParser, FormParser]
+
+    def post(self, request, pk, *args, **kwargs):
+        movie = get_object_or_404(Movie, pk=pk)
+        serializer = MovieImageUploadSerializer(movie,
+                                                data=request.data,
+                                                partial=True)
+        if serializer.is_valid():
+            serializer.save()
```
dxrrkwm, line 198: i guess it's better to use raise_exception=True here

cinema_service/settings.py
```diff
@@ -20,9 +20,8 @@
 # See https://docs.djangoproject.com/en/4.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = (
+SECRET_KEY =\
```
dxrrkwm, line 23: ?

dxrrkwm (approved):
good overall but check comments please :) 

also remove pushed images and fix your line separators to not replace every file in the project

PR #740, opened 2025-01-27
Solution

zave52:


media/upload/movies/sample-movie_-_147fa6d2-e158-4f05-8698-87f55b965cf3.jpg
SerhiyTo, line 1: Add the files generated as a result of the tests to the git log. That is, as far as I understood, these are all files that have `sample-movie` files at the beginning

cinema/views.py
```diff
@@ -101,8 +105,28 @@ def get_serializer_class(self):
         if self.action == "retrieve":
             return MovieDetailSerializer
 
+        if self.action == "upload_image":
+            return MovieImageSerializer
+
         return MovieSerializer
 
+    @action(
```
SerhiyTo, line 112: I think it's not the best practice to use ViewSet and GenericViewSet at the same time

SerhiyTo (changes requested):

media/upload/movies/sample-movie_-_425d10ac-78f2-49a4-8c3c-a94bac8372a0.jpg
SerhiyTo, line 1: The images created as a result of testing are still available in the repository :(

SerhiyTo (changes requested):

vasylhnatiuk (approved):

PR #755, opened 2025-02-13
Solution

Chukhno-Maksym:


user/serializers.py
```diff
@@ -1,24 +1,61 @@
...
+
         if password:
             user.set_password(password)
             user.save()
 
         return user
+
+
+class AuthTokenSerializer(serializers.Serializer):
+    email = serializers.CharField(label=_("Email"), write_only=True)
```
4ikkamony, line 33: Why not use serializers.EmailField, it's a bit more user friendly

4ikkamony (approved):
Very nice!

PR #757, opened 2025-02-13
Solution

Dogober:


valerii-kashpur (approved):
Images should not be pushed to git. they autogenerated just for tests and weren't deleted automatically by test clear

PR #765, opened 2025-02-14
Solution

bodiakof:


cinema/models.py
```diff
@@ -41,11 +41,16 @@ class Movie(models.Model):
     duration = models.IntegerField()
     genres = models.ManyToManyField(Genre)
     actors = models.ManyToManyField(Actor)
+    image = models.ImageField(
+        upload_to="media/images/",
```
Nattalli: nope, you had to write a function that does something like that;
 input: `image.jpg`
upload_to="upload/movies" --> 
output: `media/upload/movies/image-SOME-RANDOM_SYBMOLS.jpg`

cinema/views.py
```diff
@@ -173,3 +179,30 @@ def get_serializer_class(self):
 
     def perform_create(self, serializer):
         serializer.save(user=self.request.user)
+
+
+class MovieImageUploadView(APIView):
```
Nattalli, line 184: suggest you write own custom action
```python
 @action(
        methods=["POST"],
        detail=True,
        url_path="upload-image"
    )
```

Nattalli (changes requested):

Nattalli (approved):

PR #778, opened 2025-02-17
solution

kuinorymes:


Bleit1 (approved):
fix flake8, but anyway looks good

PR #783, opened 2025-02-20
Solution

Gleb1395:


media/uploads/movies/inception-7412a0b8-d61a-4df9-92e9-259ee0e71db1.jpg
vsmutok, line 1: Please  remove media from PR

cinema/views.py
```diff
@@ -101,8 +103,25 @@ def get_serializer_class(self):
...
+    @action(
+        methods=["POST"],
+        detail=True,
+        permission_classes=[IsAdminOrIfAuthenticatedReadOnly],
+        serializer_class=MovieImageSerializer,
+    )
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+        if serializer.is_valid():
```
vsmutok: 1. Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

Good example:
```python
if request.method == 'POST':
    serializer = MovieSerializer(data=request.data)
    serializer.save()
    return Response(serializer.data, status=status.HTTP_201_CREATED)
```

Bad example:
```python
if request.method == 'POST':
    serializer = MovieSerializer(data=request.data)
    if serializer.is_valid():
        serializer.save()
        return Response(serializer.data, status=status.HTTP_201_CREATED)
    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #793, opened 2025-03-17
Solutions

JohnyJDX:


cinema/models.py
```diff
@@ -35,12 +41,21 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(instance: "Movie", filename: str) -> pathlib.Path:
+    filename = (
+        f"{slugify(instance.title)}-{uuid.uuid4()}"
+        + pathlib.Path((filename)).suffix
+    )
+    return pathlib.Path("upload/movies") / pathlib.Path(filename)
```
Nattalli: you don't need to use `pathlib.Path(filename)`
It can be simplified to `pathlib.Path("uploads/movies") / filename`

Nattalli (changes requested):

Nattalli (approved):

PR #828, opened 2025-04-29
completed

BGBroth:


BGBroth:
@mateacademy-ai-mentor

i changed the serializer class but your second suggestion is wrong cause i have a separate serializer for image upload called MovieImageSerializer

PR #848, opened 2025-05-06
Solution

Tolik-code:


AnyoneClown (changes requested):
Add `media` folder to `.gitignore` and delete all files from this folder on remote repo

PR #861, opened 2025-05-18
'Solution'

Julia15031988:


cinema/models.py
```diff
@@ -35,12 +39,20 @@ def full_name(self):
         return f"{self.first_name} {self.last_name}"
 
 
+def movie_image_path(instance: "Movie", filename: str) -> pathlib.Path:
+    filename = f"{slugify(instance.title)}-{uuid.uuid4()}" \
```
viktoria-rybenchuk: you should use parentheses instead of backslash

viktoria-rybenchuk (changes requested):

vsmutok (approved):
Great!

PR #925, opened 2025-09-05
create adding image with auth, also change username to email

SerhiiBasok:


.github/workflows/test.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
+          pip install Pillow
```
AnyoneClown, line 26: Just add Pillow into `requirements.txt`. Don't change CI/CD

AnyoneClown (approved):
LGTM, small issue in CI/CD

PR #932, opened 2025-09-08
image upload and login via email

Vskesha:


Otst0y (approved):

Vskesha:
@mateacademy-ai-mentor Everything was fixed

Vskesha:
@mateacademy-ai-mentor I will not change any migration as they are generated automatically by django. Also field username can be in user model and serializer, but I made it not necessary and now required field for authentication is email field. This works properly.

PR #939, opened 2025-09-09
Solution

Artcrafterrra:


user/serializers.py
```diff
@@ -22,3 +22,28 @@ def update(self, instance, validated_data):
...
+        trim_whitespace=False,
+        write_only=True,
+    )
+
+    def validate(self, attrs):
+        email = attrs.get("email")
+        password = attrs.get("password")
+        user = authenticate(
+            request=self.context.get("request"),
+            username=email,
```
MakovM: Use email=email instead of username=email when calling authenticate, this causes authentication to always fail

MakovM (changes requested):
Your solution must pass all tests.

user/serializers.py
Artcrafterrra: THX fixed

MakovM (approved):
LGTM

cinema/views.py
```diff
@@ -101,8 +104,22 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        movie = self.get_object()
+        serializer = self.get_serializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
AnyoneClown: Refactor using `raise_exceptions=True`

AnyoneClown (changes requested):

L1nk3rrr (approved):
Well done!

PR #948, opened 2025-09-10
Solution

Irahromska:


Otst0y (changes requested):
You need to alter your db (delete all username entries) in order to pass tests.

PR #1020, opened 2026-01-04
Solution

vakt159:


cinema/views.py
```diff
@@ -101,23 +110,36 @@ def get_serializer_class(self):
...
+    @action(methods=["POST"], detail=True, url_path="upload-image")
+    def upload_image(self, request, pk=None):
+        item = self.get_object()
+        serializer = self.get_serializer(item, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
AngryAndrii, line 135: There is better way to write this condition - `serializer.is_valid(raise_exception=True)` - we can avoid rewriting return `Response(...`

cinema/views.py
```diff
@@ -21,7 +30,7 @@
     MovieSessionDetailSerializer,
     MovieListSerializer,
     OrderSerializer,
-    OrderListSerializer,
+    OrderListSerializer, MovieImageSerializer,
```
AngryAndrii: better move this to the next line

AngryAndrii (approved):
Good work! I didn't see any critical mistakes

